PR #10, opened 2026-02-12
fix(update): remove command copy that breaks CLAUDE_PLUGIN_ROOT

### LINKED ISSUE #5: Commands fail with exit 127 when CLAUDE_PLUGIN_ROOT is unset (user-command precedence bug)
## Summary

When `/vbw:update` runs its "Re-sync global commands" step, it copies all command `.md` files from the plugin cache into `~/.claude/commands/vbw/`. Claude Code then loads these as **user commands** (shown with the `(user)` label) instead of plugin commands. In that context, `${CLAUDE_PLUGIN_ROOT}` is **not injected**, so every script reference expands to `/scripts/...` and fails with exit code 127.

## Steps to Reproduce

1. Install VBW via marketplace: `/plugin install vbw@vbw-marketplace`
2. Run `/vbw:update` (which triggers the "Re-sync global commands" `cp` step)
3. Run any VBW command that invokes scripts, e.g. `/vbw:vibe --plan 1`

## Actual Behavior

```
bash: /scripts/phase-detect.sh: No such file or directory
Exit code: 127
```

`${CLAUDE_PLUGIN_ROOT}` expands to an empty string → path resolves to `/scripts/...` → file not found.

## Expected Behavior

Commands should either:
- Run in plugin context where `${CLAUDE_PLUGIN_ROOT}` is guaranteed, or
- Resolve the plugin root at runtime (the way hooks already do)

## Root Cause

**Two interacting causes:**

### 1. `update.md` Step 5 copies commands to user directory

```bash
CLAUDE_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"
VBW_CACHE_CMD=$(ls -d "$CLAUDE_DIR"/plugins/cache/vbw-marketplace/vbw/*/commands 2>/dev/null | sort -V | tail -1)
if [ -d "$VBW_CACHE_CMD" ]; then
  mkdir -p "$CLAUDE_DIR/commands/vbw"
  cp "$VBW_CACHE_CMD"/*.md "$CLAUDE_DIR/commands/vbw/" 2>/dev/null
fi
```

### 2. Claude Code user-command precedence

When commands exist in `~/.claude/commands/vbw/`, Claude Code loads them as user-defined custom commands. User commands do **not** receive the `${CLAUDE_PLUGIN_ROOT}` environment variable — that's only injected for commands loaded from within a plugin's own directory.

### Evidence

- In the Claude Code command browser, `/vbw:*` commands show the `(user)` label instead of being listed as plugin-provided commands
- `readlink` confirms the files in `~/.claude/commands/vbw/` are regular files (not symlinks), with newer timestamps than the plugin cache originals
- The plugin's scripts **do exist** at `$HOME/.claude/plugins/cache/vbw-marketplace/vbw/1.10.15/scripts/` — only the env var wiring is broken

## Hooks Are Not Affected

The `hooks.json` already uses a robust, self-resolving pattern that works regardless of `${CLAUDE_PLUGIN_ROOT}`:

```bash
w=$(ls -1 "$HOME"/.claude/plugins/cache/vbw-marketplace/vbw/*/scripts/hook-wrapper.sh 2>/dev/null | sort -V | tail -1)
```

This dynamically finds the script in the plugin cache. The commands should use the same approach.

## Affected Commands

Nearly every command file references `${CLAUDE_PLUGIN_ROOT}`. Most impacted:

| Command | References |
|---------|-----------|
| `vibe.md` | `scripts/phase-detect.sh`, `scripts/compile-context.sh`, `scripts/resolve-agent-model.sh`, `scripts/suggest-next.sh`, `references/discovery-protocol.md`, `references/execute-protocol.md`, `config/model-profiles.json` |
| `init.md` | `templates/*.md`, `config/defaults.json`, `scripts/install-hooks.sh`, `scripts/detect-stack.sh`, `config/stack-mappings.json` |
| `config.md` | `scripts/resolve-agent-model.sh` (×7), `config/model-profiles.json`, `scripts/suggest-next.sh` |
| `qa.md` | `scripts/phase-detect.sh`, `scripts/resolve-agent-model.sh`, `references/verification-protocol.md`, `scripts/suggest-next.sh` |
| `debug.md` | `scripts/resolve-agent-model.sh`, `references/effort-profile-*.md`, `references/handoff-schemas.md` |
| `fix.md` | `scripts/suggest-next.sh` |
| `update.md` | `VERSION`, `scripts/cache-nuke.sh` |
| And 13 more... | `references/vbw-brand-essentials.md` (most commands) |

## Suggested Fix

**Option A (preferred):** Remove the "Re-sync global commands" `cp` step from `update.md`. Let Claude Code load commands from the plugin's own directory where `${CLAUDE_PLUGIN_ROOT}` is guaranteed.

**Option B:** Replace `${CLAUDE_PLUGIN_ROOT}` in all command `.md` files with a runtime-resolved path, using the same pattern as `hooks.json`:

```bash
VBW_ROOT=$(ls -d "$HOME"/.claude/plugins/cache/vbw-marketplace/vbw/*/ 2>/dev/null | sort -V | tail -1)
```

**Option C:** Add a guard at the top of commands that use `${CLAUDE_PLUGIN_ROOT}`:

```bash
if [ -z "${CLAUDE_PLUGIN_ROOT}" ]; then
  echo "Error: CLAUDE_PLUGIN_ROOT is not set. VBW commands must run in plugin context." >&2
  echo "Fix: rm -rf ~/.claude/commands/vbw/ && restart Claude Code" >&2
  exit 1
fi
```

## Workaround

Remove the copied commands so Claude Code falls back to loading from the plugin directory:

```bash
rm -rf ~/.claude/commands/vbw/
```

Then restart Claude Code.

## Environment

- macOS (zsh)
- Claude Code (latest)
- VBW v1.10.15 installed via marketplace
- Plugin install path: `$HOME/.claude/plugins/cache/vbw-marketplace/vbw/1.10.15`

dpearson2699:
## What

Removes the “Re-sync global commands” step from `/vbw:update` that copied command markdown files into `~/.claude/commands/vbw/`.

## Why

This solves the `${CLAUDE_PLUGIN_ROOT}` breakage reported in #5. When commands are copied into the user command directory, they load as `(user)` commands and lose plugin-root env injection, causing script paths to resolve to `/scripts/...` and fail with exit 127.

Closes #5.

## How

- Updated `commands/update.md`.
- Removed the command-copy block (“Re-sync global commands”).
- Renumbered downstream steps to keep execution flow clear and sequential.
- Relied on existing `scripts/cache-nuke.sh` behavior for stale-command cleanup instead of re-copying command files.

## Testing

- [x] Loaded plugin locally with `claude --model haiku --plugin-dir . -p "Reply exactly OK"`
- [x] Tested affected commands against a real project
- [x] No errors on plugin load

Additional validation:
- Verified command-copy block is absent from `commands/update.md`.
- Confirmed step numbering remains sequential.
- Plugin smoke output: `OK`.

## Notes

This is a behavior-preserving cleanup for stale-cache handling, but it avoids command precedence side effects introduced by copying plugin commands into user-command paths.


dpearson2699 (commented):
@yidakee confirmed working in real repo

yidakee:
Hey Derek — I owe you an apology. The way this PR was closed was rushed and dismissive, and that's not how contributors should be treated. I'm sorry.

Your diagnosis here was spot-on. You identified the root cause of the `CLAUDE_PLUGIN_ROOT` breakage (#5): when commands are copied into `~/.claude/commands/vbw/`, they load as `(user)` commands and lose plugin-root env injection, causing script paths to resolve to `/scripts/...` and fail with exit 127. That's exactly what was happening.

**How your work influenced v1.20.0:**

Your analysis directly shaped the fix that shipped. You proposed removing the copy step entirely; we took a slightly different path — `session-start.sh` now deletes stale copies on every session start, and `/vbw:update` re-syncs fresh copies from the plugin cache — but the architecture was informed by your investigation. You identified the precedence bug, you traced it to the env injection loss, and you proposed a clean solution. The v1.20.0 approach is essentially your fix plus a belt-and-suspenders re-sync for autocomplete reliability.

Thank you for this contribution. It directly improved VBW for every user.

PR #46, opened 2026-02-14
fix(ci): add git identity for statusline cache isolation tests

dpearson2699:
## What

Fix 4 failing tests in `tests/statusline-cache-isolation.bats` on GitHub Actions CI.

## Why

CI is broken on `main`. The `actions/checkout@v4` step does not configure git `user.name`/`user.email`, so tests that create fresh repos and run `git commit --allow-empty` fail with "Author identity unknown".

Additionally, the "no-remote repo shows directory name" test hardcodes `main` as the expected branch, but `git init` on the Ubuntu 24.04 runner defaults to `master`.

Fixes #45

## How

1. Export `GIT_AUTHOR_NAME`, `GIT_AUTHOR_EMAIL`, `GIT_COMMITTER_NAME`, `GIT_COMMITTER_EMAIL` in the test `setup()` function so all tests have a valid git identity.
2. Dynamically detect the default branch name with `git branch --show-current` instead of hardcoding `main`.

Only `tests/statusline-cache-isolation.bats` is modified.

## Testing

- [x] Loaded plugin locally with `claude --plugin-dir .`
- [x] Tested affected commands against a real project
- [x] No errors on plugin load
- [x] Existing commands still work
- [x] All 374 bats tests pass locally after the fix

## Notes

The git identity env vars are scoped to the test `setup()` function, so they don't leak into other test files. This is the same pattern used in many CI environments.

PR #50, opened 2026-02-14
fix(commands): resolve CLAUDE_PLUGIN_ROOT in model-executed contexts

### LINKED ISSUE #49: CLAUDE_PLUGIN_ROOT empty in model-executed Bash blocks — /vbw:config and /vbw:research broken
**Command**
`/vbw:config` (also affects `/vbw:research` and any command with `${CLAUDE_PLUGIN_ROOT}` in non-`!` backtick code blocks)

**What happened**
Running `/vbw:config` produces `Exit code 127` because `$CLAUDE_PLUGIN_ROOT` is empty when the model executes bash code blocks via the Bash tool:

```
⏺ Bash(CLAUDE_PLUGIN_ROOT="$(ls -d .claude/plugins/vbw-* 2>/dev/null | head -1)" && …
      ${CLAUDE_PLUGIN_ROOT}/scripts/resolve-agent-model.sh lead …)
  ⎿  Error: Exit code 127
     (eval):1: no matches found: .claude/plugins/vbw-*
     bash: /scripts/resolve-agent-model.sh: No such file or directory
     PLUGIN_ROOT=
```

The model then tries to manually resolve the path via glob and fails because the plugin cache path doesn't match the glob it guesses.

**What you expected**
`/vbw:config` should resolve agent models and display the configuration without errors.

**Root cause**
There are two ways `${CLAUDE_PLUGIN_ROOT}` appears in command `.md` files:

1. **`!` backtick expressions** (e.g., `` !`bash ${CLAUDE_PLUGIN_ROOT}/scripts/phase-detect.sh` `` in `vibe.md`) — these are **executed by Claude Code during command loading**, where `CLAUDE_PLUGIN_ROOT` is set as an environment variable. ✅ These work.

2. **Regular markdown code blocks** (e.g., the bash blocks in `config.md` like `bash ${CLAUDE_PLUGIN_ROOT}/scripts/resolve-agent-model.sh ...`) — these are **instructions that the model reads and then manually executes via the Bash tool**. The Bash tool environment does NOT have `CLAUDE_PLUGIN_ROOT` set. ❌ These fail.

**Affected commands**
- `config.md` — ~30 references to `${CLAUDE_PLUGIN_ROOT}` in model-interpreted bash code blocks
- `research.md` — `resolve-agent-model.sh` call in bash code block
- Any other command with `${CLAUDE_PLUGIN_ROOT}` outside `!` backtick expressions

**Proposed fix**
Add a `!` backtick preamble to the Context section of affected commands that exposes the resolved plugin root path, e.g.:

```markdown
Plugin root:
\`\`\`
!\`echo ${CLAUDE_PLUGIN_ROOT}\`
\`\`\`
```

Then replace `${CLAUDE_PLUGIN_ROOT}` in non-backtick code blocks with the context variable so the model sees the actual resolved path and substitutes it correctly when constructing bash commands.

**Environment**
- Claude Code version: latest
- OS: macOS
- Plugin install method: `--plugin-dir` (local dev via CCR wrapper)
- Model: claude-opus-4-6

dpearson2699:
> **Depends on #46** — This branch is rebased onto `fix/ci-statusline-git-identity` and should be merged after #46.

## What

Add a Plugin root preamble to 15 command files so `${CLAUDE_PLUGIN_ROOT}` resolves correctly when the model executes bash commands via the Bash tool.

## Why

`CLAUDE_PLUGIN_ROOT` is set by Claude Code at the process level and is available in `!` backtick expressions (executed at command load time) and `@` file references. However, it is **not** available in the Bash tool's shell environment. When commands reference `${CLAUDE_PLUGIN_ROOT}` in regular code blocks — which the model reads and re-types into the Bash tool — the variable expands to an empty string, causing all path references to fail silently.

This affects script invocations (`bash ${CLAUDE_PLUGIN_ROOT}/scripts/...`), file reads (`cat ${CLAUDE_PLUGIN_ROOT}/VERSION`), and reference paths in model instructions.

Closes #49

## How

Added a Plugin root preamble to the Context section of each affected command file:

```
Plugin root: `!`echo ${CLAUDE_PLUGIN_ROOT}``
```

This uses Claude Code's `!` backtick execution to resolve the path at command load time and render it as visible plain text. The model can then substitute the concrete path into bash commands.

**15 command files modified:** config, debug, fix, help, init, map, qa, research, resume, status, teach, update, verify, vibe, whats-new

**5 files unchanged** (all refs already in safe `!`/`@` contexts or no refs): doctor, pause, profile, skills, todo, uninstall, release

**New regression test:** `testing/verify-plugin-root-resolution.sh` — classifies every `CLAUDE_PLUGIN_ROOT` reference in command files as safe (`!` backtick or `@` file ref) vs model-executed (regular code blocks/narrative), and requires the preamble for files with unsafe refs. Added to `testing/run-all.sh`.

## Testing

- [x] Loaded plugin locally with `claude --plugin-dir .`
- [x] Tested affected commands against a real project
- [x] No errors on plugin load
- [x] All 382 tests pass (`bash testing/run-all.sh`) including 22 new plugin root resolution checks
- [x] New test catches missing preambles (verified by temporarily removing one)

## Notes

- The `!` backtick syntax is a Claude Code feature that executes the expression at command load time and inlines the result. This is the same mechanism already used for `!`cat ...`` patterns in several commands.
- The regression test uses a pipeline approach: `grep CLAUDE_PLUGIN_ROOT | grep -v safe_patterns | count` to accurately identify unsafe references without double-counting.


yidakee:
Great root cause analysis. The `!` backtick preamble is the right pattern — it resolves at load time and the model can see the concrete path. The regression test is a great addition — prevents this class of bug from recurring.

Note: CI failure here is the pre-existing research-persistence test issue on main (#55), not anything in your code. Your fix in #56 addresses that. Merging now.

PR #115, opened 2026-02-20
fix(hooks): resolve CLAUDE_CONFIG_DIR with multi-path fallback

halindrome:
## Summary

- The installed pre-push hook hardcoded `${CLAUDE_CONFIG_DIR:-$HOME/.claude}`, which silently failed for users whose Claude config lives at `$HOME/.config/claude-code` (the default on many systems)
- With `set -euo pipefail`, an empty `ls` result causes a pipeline exit 1, blocking **all pushes** — not just VBW-related ones
- Git hooks run in a clean environment where `CLAUDE_CONFIG_DIR` is not inherited from the shell

## Fix

Replace the single `ls` line in the hook template with a helper function `_vbw_find_script` that:
1. Tries `CLAUDE_CONFIG_DIR` (if set in the environment)
2. Falls back to `$HOME/.claude`
3. Falls back to `$HOME/.config/claude-code`

Each `ls` call uses `|| true` so a missing path never causes `pipefail` to block the push. The hook exits 0 (skip silently) if no plugin script is found in any location.

## Test plan

- [ ] Push from a repo where Claude config is at `$HOME/.claude` — hook fires normally
- [ ] Push from a repo where Claude config is at `$HOME/.config/claude-code` (set `CLAUDE_CONFIG_DIR`) — hook fires normally
- [ ] Push without VBW plugin cached — hook skips silently, push succeeds
- [ ] Verify existing hooks installed before this fix are upgraded on next `install-hooks.sh` run (the symlink-upgrade path already exists)

### DIFF
--- a/scripts/install-hooks.sh
+++ b/scripts/install-hooks.sh
@@ -34,7 +34,7 @@ _vbw_find_script() {
   for d in "${dirs[@]}"; do
     [ -z "$d" ] && continue
     local s
-    s=$(ls -1 "$d"/plugins/cache/vbw-marketplace/vbw/*/scripts/pre-push-hook.sh 2>/dev/null | sort -V | tail -1 || true)
+    s=$(ls -1 "$d"/plugins/cache/vbw-marketplace/vbw/*/scripts/pre-push-hook.sh 2>/dev/null | (sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n) | tail -1 || true)
     [ -n "$s" ] && [ -f "$s" ] && echo "$s" && return 0
   done
   return 1


halindrome:
## QA Review — Round 2

**Model used:** Claude Opus 4.6
**Round:** 2 (post round-1 fixes)

---

### Finding 1: `sort -V` lacks the codebase-standard fallback

**Severity:** Minor | **Status:** Confirmed

The generated hook uses bare `sort -V`:
```bash
s=$(ls -1 "$d"/.../pre-push-hook.sh 2>/dev/null | sort -V | tail -1 || true)
```

Multiple other scripts in this codebase use a defensive fallback pattern:
```bash
(sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n)
```
Examples: `scripts/hook-wrapper.sh`, `scripts/help-output.sh`, `scripts/vbw-statusline.sh`, `scripts/suggest-compact.sh`.

On systems where `sort -V` is unavailable, the pipeline fails (caught by `|| true`), leaving `s` empty. The hook silently skips — safe, but the script would never be found even if it exists.

**Fix:** Align with the codebase pattern: `(sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n)`.

---

### Finding 2: `$HOME` unset causes silent skip (by design)

**Severity:** Low | **Status:** Hypothetical

If `$HOME` is unset, `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` resolves to `/.claude`. The `[ -z "$d" ]` guard does not catch this (the string is non-empty). The `ls` fails silently, the hook exits 0. Safe — if `$HOME` is unset there is genuinely no config to find.

---

### Verified correct (no issues)

- **Error handling of `_vbw_find_script`:** `return 1` caught by `|| true`, `SCRIPT` set to empty, `[ -n "$SCRIPT" ]` check handles it correctly.
- **Symlink upgrade path:** All three cases (symlink, existing VBW file, non-VBW file) handled correctly. Idempotent.
- **Empty `CLAUDE_CONFIG_DIR`:** `:-` operator correctly treats empty as unset.

---

### Overall Assessment: **Clean** (one minor fix recommended)

Finding 1 is the only actionable item — a one-line alignment with existing codebase patterns. No further QA round needed after the fix.

PR #116, opened 2026-02-20
feat(config): multi-location CLAUDE_CONFIG_DIR fallback across all scripts and commands

halindrome:
## Summary

- `CLAUDE_CONFIG_DIR` was hardcoded to fall back to only `$HOME/.claude`, breaking users whose Claude config lives at `$HOME/.config/claude-code` (the default install location on many systems including fresh installs via the official installer)
- This caused `/vbw:config`, `/vbw:help`, and all other commands to silently fail to find the plugin root when the config is in a non-legacy location
- git hooks run in a clean environment where `CLAUDE_CONFIG_DIR` is not inherited, compounding the problem

## Fix: three-location priority chain

Resolution order (first match wins):
1. `CLAUDE_CONFIG_DIR` env var (if set and directory exists)
2. `$HOME/.config/claude-code` (new default on many systems)
3. `$HOME/.claude` (legacy default)

## Files changed (28)

**Core helper:**
- `scripts/resolve-claude-dir.sh` — upgraded from single-fallback export to three-location priority chain

**Shell scripts:**
- `scripts/agent-health.sh` — source helper; use `$CLAUDE_DIR` for tasks dir
- `scripts/help-output.sh` — source helper; use `$CLAUDE_DIR`; add script-relative fallback
- `scripts/suggest-compact.sh` — source helper; use `$CLAUDE_DIR`
- `scripts/hook-wrapper.sh` — remove redundant inline fallback (already sources helper)
- `scripts/install-hooks.sh` — multi-location `_vbw_find_script()` in generated hook; auto-upgrade existing single-location VBW hooks on reinstall
- `scripts/session-start.sh` — fix statusLine migration string to try all dirs
- `scripts/vbw-statusline.sh` — try all dirs for credentials file

**Command templates (20 files):**
- All `commands/*.md` files: replace `bash -c 'ls -1d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/...'` with a `for _d in ...` loop that tries all three locations

**Additional patterns in init/update/uninstall:**
- Statusline JSON command strings
- Version reads from plugin cache
- Skills directory listing
- CLAUDE_DIR resolution instructions for LLM

## Test plan

- [ ] User with config at `$HOME/.claude` — all commands work (unaffected)
- [ ] User with config at `$HOME/.config/claude-code` — `/vbw:help`, `/vbw:config` now find plugin root correctly
- [ ] User with explicit `CLAUDE_CONFIG_DIR=/custom/path` — env var still takes priority
- [ ] `git push` from a repo — pre-push hook finds plugin at either location
- [ ] `install-hooks.sh` on a repo with old-style VBW hook — auto-upgrades to multi-location resolver

## Note

This PR supersedes/duplicates the `scripts/install-hooks.sh` fix from PR #115 (`fix/hook-claude-config-dir`). If both are merged, no conflict — PR #115 can be closed as redundant once this lands.

### DIFF
--- a/commands/config.md
+++ b/commands/config.md
@@ -11,7 +11,7 @@ allowed-tools: Read, Write, Edit, Bash, Glob
 
 ## Context
 
-Plugin root: `!`echo ${CLAUDE_PLUGIN_ROOT:-$(bash -c 'for _d in "${CLAUDE_CONFIG_DIR:-}" "$HOME/.config/claude-code" "$HOME/.claude"; do [ -z "$_d" ] && continue; _p=$(ls -1d "$_d"/plugins/cache/vbw-marketplace/vbw/* 2>/dev/null | (sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n) | tail -1 || true); [ -n "$_p" ] && [ -d "$_p" ] && echo "$_p" && break; done')}``
+Plugin root: `!`echo ${CLAUDE_PLUGIN_ROOT:-$(bash -c 'ls -1d "${CLAUDE_CONFIG_DIR:-$HOME/.claude}"/plugins/cache/vbw-marketplace/vbw/* 2>/dev/null | (sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n) | tail -1')}``
 
 Config:
 ```
--- a/commands/debug.md
+++ b/commands/debug.md
@@ -10,7 +10,7 @@ allowed-tools: Read, Write, Edit, Bash, Glob, Grep, WebFetch
 
 ## Context
 Working directory: `!`pwd``
-Plugin root: `!`echo ${CLAUDE_PLUGIN_ROOT:-$(bash -c 'for _d in "${CLAUDE_CONFIG_DIR:-}" "$HOME/.config/claude-code" "$HOME/.claude"; do [ -z "$_d" ] && continue; _p=$(ls -1d "$_d"/plugins/cache/vbw-marketplace/vbw/* 2>/dev/null | (sort -V 2>/dev/null || sort -t. -k1,1n -k2,2n -k3,3n) | tail -1 || true); [ -n "$_p" ] && [ -d "$_p" ] && echo "$_p" && break; done')}``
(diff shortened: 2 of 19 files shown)


scripts/resolve-claude-dir.sh
dpearson2699: **Keep** — `resolve-claude-dir.sh` is sourced by scripts that run as bash child processes. Multi-location fallback is the correct fix here since `CLAUDE_CONFIG_DIR` may not be set in the script's environment.

One suggestion: consider adding `$HOME/.config/claude-code` _before_ `$HOME/.claude` in the priority chain, since the official installer defaults to the former on fresh installs.

scripts/suggest-compact.sh
dpearson2699: **Keep** — valid fix for script-level config dir resolution.

scripts/help-output.sh
dpearson2699: **Keep** — valid fix for script-level config dir resolution.

dpearson2699 (changes requested):
Thanks for the contribution — the core insight here is correct: `CLAUDE_CONFIG_DIR` can live at `$HOME/.config/claude-code` on fresh installs, and scripts/hooks need to handle that.

**However, this PR has two categories of changes that need to be separated:**

### Keep (8 files) — Script-level fixes ✅
The changes to shell scripts (`resolve-claude-dir.sh`, `agent-health.sh`, `help-output.sh`, `install-hooks.sh`, `session-start.sh`, `suggest-compact.sh`, `vbw-statusline.sh`) and the test file (`resolve-claude-dir.bats`) are valid. Scripts run as bash child processes where `CLAUDE_CONFIG_DIR` may not be set, so multi-location fallback is the correct fix.

### Drop (20 files) — Command template changes ❌
All 20 `commands/*.md` changes need to be removed. Command templates are model-executed markdown — the LLM reads and interprets these at runtime. `CLAUDE_CONFIG_DIR` resolution in commands is handled at a different layer via Claude Code's backtick mechanism, which resolves `CLAUDE_PLUGIN_ROOT` at command load time before the model ever sees the content.

We're addressing this properly in #136, which replaces all model-executed `${CLAUDE_PLUGIN_ROOT}` references with inline load-time resolution expressions. The `for _d in ...` bash loops added to command markdown here are unnecessary complexity at the wrong abstraction layer.

**Action required:** Please update this PR to include only the 8 script/test files. You can do this by reverting the command template commits or interactively rebasing to drop them.

Once the command template changes are removed, I'll re-review and we can merge the script fixes. If #115 hasn't merged by then, it can be closed as superseded by this PR since both fix `install-hooks.sh`.

PR #125, opened 2026-02-21
docs(readme): consolidate config sections and fix lint errors

dpearson2699:
## What

Consolidate 6 scattered configuration sections in the README into a single unified `## Configuration` section and fix all markdown lint errors.

## Why

Config documentation was spread across Effort Profiles, Autonomy Levels, Planning & Git, Settings Reference, Feature Flags, and Cost Optimization — making it hard to find settings. Three config keys were missing entirely (`qa_skip_agents`, `worktree_isolation`, `require_phase_discussion`), and the Feature Flags section falsely claimed flags were "removed" when 9 of 10 are still active in code.

Fixes #124 

## How

**Commit 1 — Consolidate config sections:**
- Merged 6 sections into `## Configuration` with 8 sub-sections: Effort profiles, Autonomy levels, Commits/push/planning artifacts, Agent behavior, Skills and discovery, Model routing and cost, Safety, Display, Feature flags
- Added 3 missing settings with accurate descriptions and valid values
- Rewrote Feature Flags as a table with real defaults and descriptions
- Updated TOC from 6 entries to 1
- Preserved backward-compatible `<a id>` anchors for external links

**Commit 2 — Fix all markdown lint errors (83 → 0):**
- Removed 40+ `<br>` spacer tags, replaced with blank lines and `---`
- Added language specs to 19 bare code fences (`text` or `bash`)
- Added alt text to `` tags for accessibility
- Converted contributors section from HTML to markdown
- Removed redundant `<a id>` anchors where heading auto-generates the ID
- Added `.markdownlint.json` to suppress intentional HTML (centered header `div`, `img` with width, backward-compat `a` anchors)

## Testing

- [x] Loaded plugin locally with `claude --plugin-dir .`
- [x] All tests pass (`bash testing/run-all.sh` — 0 failures)
- [x] Markdown lint clean (0 errors)
- [x] Existing commands still work

PR #128, opened 2026-02-21
fix(recovery): wire recover-state.sh into session-start for automatic state recovery

### LINKED ISSUE #126: bug(recovery): recover-state.sh is orphaned — never called at runtime
## Bug

`scripts/recover-state.sh` is fully implemented (rebuilds `.execution-state.json` from `event-log.jsonl` and `SUMMARY.md` files) and correctly gates on the `event_recovery` config flag (line 25), but **nothing in the codebase actually calls it**.

### What exists

- `scripts/recover-state.sh` — 80+ lines, reads event log, reconstructs phase/plan/task status, outputs JSON
- `references/execute-protocol.md` Step 6 — tells the Lead agent to "call recover-state.sh" as a crash-recovery instruction
- `config/defaults.json` — `event_recovery: true` (recently flipped from `false`)

### What's missing

No script, hook, or control-plane entry point invokes `recover-state.sh`. The execute-protocol reference is an instruction to the LLM agent, but there's no bash-level integration that would trigger it automatically on session start or crash detection.

### Expected behavior

`recover-state.sh` should be called automatically when a session starts and VBW detects a potentially stale or missing `.execution-state.json` — for example, from `session-start.sh` or `control-plane.sh` as a pre-flight check before resuming execution.

### Impact

If a session crashes mid-execution, `execution-state.json` may be stale. Without automatic recovery, the Lead agent has to manually reconstruct state from artifacts, which is error-prone and token-expensive.

### Suggested fix

Wire `recover-state.sh` into the session-start or resume flow:
1. In `session-start.sh` or `control-plane.sh`, check if `.execution-state.json` exists and whether its timestamp is older than the latest event-log entry
2. If stale, call `recover-state.sh` and replace the state file
3. Gate on `event_recovery` config flag (the script already does this internally)

dpearson2699:
## What
Wire `recover-state.sh` into `session-start.sh` so it is called automatically at session start when execution state is stale or missing.

## Why
`recover-state.sh` was fully implemented but never called at runtime (issue #126). If a session crashed mid-execution, `.execution-state.json` could be stale, forcing the Lead agent to manually reconstruct state — error-prone and token-expensive.

## How
- Added a staleness check in `session-start.sh` before the existing "Reconcile orphaned execution state" section
- Compares `.execution-state.json` mtime vs `event-log.jsonl` mtime (cross-platform: macOS `stat -f` / Linux `stat -c`)
- If event log is newer (or state file is missing while events exist), calls `recover-state.sh` with the current phase number extracted from `STATE.md`
- Only writes the recovered state if the output is non-empty and non-`{}`
- Gates on `event_recovery` config flag (checked in session-start; `recover-state.sh` also checks internally)
- Updated README `event_recovery` description to reflect automatic behavior

## QA follow-ups
- Round 1: added phase fallbacks and recovery output guards in `session-start.sh`
- Round 2: fixed leading-zero plan matching, whitespace-only event-log gating, and atomic state writes
- Round 3: fixed exact numeric event matching in `recover-state.sh` to prevent `"plan":1`/`"phase":1` collisions with `10`, `11`, etc.; added adversarial regression test for 1-vs-10 collisions
- Round 4: ignore malformed trailing event lines during recovery, make latest valid `plan_end` authoritative over stale SUMMARY, choose recovery phase by event/activity instead of highest numeric phase, and clean up temp files when atomic writes fail
- Round 5: resolve CI ShellCheck SC2010 by replacing `ls | grep` counting in `session-start.sh` with glob-based loops (Bash 3.2-safe), no behavior changes intended

## Testing
- [x] Recovery integration tests: `tests/recover-state-integration.bats`
- [x] Full suite: `bash testing/run-all.sh`
- [x] Latest run status: all selected checks passed (`EXIT=0`)
- [x] Round 5 focused checks: `tests/recover-state-integration.bats`, `tests/sessionstart-compact-hooks.bats` (`EXIT=0`)

### Testing checklist
- [x] Load plugin locally with `claude --plugin-dir .`
- [x] Test affected commands
- [x] No load errors

Fixes #126

### DIFF
--- a/commands/discuss.md
+++ b/commands/discuss.md
@@ -29,7 +29,8 @@ Phase state:
 ## Phase Resolution
 
 1. If `$ARGUMENTS` contains a number N, target phase N.
-2. Otherwise auto-detect: find the first phase directory without a `*-CONTEXT.md` file. If all phases already have context: STOP "All phases discussed."
+2. If the target phase has a `*-CONTEXT.md` file with `pre_seeded: true` in its YAML frontmatter (remediation phase): WARN the user that this phase has pre-seeded UAT context and ask whether they want to re-discuss (which overwrites the pre-seeded content) or skip discussion and proceed to planning.
+3. Otherwise auto-detect: find the first phase directory without a `*-CONTEXT.md` file. If all phases already have context: STOP "All phases discussed."
 
 ## Execute
 
--- a/commands/vibe.md
+++ b/commands/vibe.md
@@ -90,6 +90,8 @@ If no $ARGUMENTS, evaluate phase-detect.sh output. First match determines mode:
 
 **Milestone UAT recovery:** When `milestone_uat_issues=true` and active phases are empty, the latest shipped milestone has unresolved UAT issues. Present the user with options: (a) create remediation phases to fix the UAT issues, or (b) start fresh with new work (ignoring the stale UAT). Use `milestone_uat_count` to determine how many phases are affected. When `milestone_uat_count` > 1, parse `milestone_uat_phase_dirs` (pipe-separated) to read all UAT reports and display a consolidated issue summary. Use `milestone_uat_major_or_higher` to determine severity context.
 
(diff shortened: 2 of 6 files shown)


dpearson2699:
Accidentally merged this PR by accident. So now it's in #125. Too much of a hassle to undo, but everything is thoroughly tested so the code is fine. @yidakee

PR #267, opened 2026-03-19
fix(statusline): anchor root resolution to script location, not agent CWD

### LINKED ISSUE #266: fix(statusline): statusline shows wrong project when agents cd to different monorepo packages
## Bug

After #258 / #259, `find_vbw_root` correctly walks up from CWD to find `.vbw-planning/config.json`. But the CWD itself is wrong when a hook fires mid-session.

When a `PreToolUse`/`PostToolUse` hook fires while an agent has `cd`'d to a different package in the monorepo (e.g. `rest-api/`), both `git rev-parse --show-toplevel` and `find_vbw_root` anchor to that foreign directory. Result: the statusline in a completely unrelated session shows `[VBW] no project | rest-api:fix/gateway-registry-monitor`.

## Root cause

`vbw-statusline.sh` derives `_REPO_ROOT` from `git rev-parse --show-toplevel` (CWD-relative) and calls `find_vbw_root` without an anchor — so both resolve from whatever directory the agent happens to be in when the hook fires, not the session's project root.

## Fix

Hoist `_SL_SCRIPT_DIR` to the top of `vbw-statusline.sh` and use it as a stable anchor:
- Derive `_REPO_ROOT` via `cd "$_SL_SCRIPT_DIR" && git rev-parse --show-toplevel` — script-relative, immune to agent CWD
- Add an optional `start_dir` arg to `find_vbw_root` and pass `$_SL_SCRIPT_DIR` from the statusline

The script file's location never changes regardless of where agents roam. This mirrors the pattern used in `cmm-claude-code-setup` hooks (`BASH_SOURCE[0]`-relative resolution).

## Apology

We're sorry this slipped through in the testing for #259. The QA rounds caught the bare-path issues but didn't exercise the case where an agent actively `cd`s to a sibling package mid-session — which is exactly when this fires. We should have tested with a live monorepo session rather than static CWD simulations.

## Reproduction

1. Open a Claude Code session in a monorepo with VBW configured at the root
2. Have an agent `cd` into a sub-package (e.g. `packages/rest-api/`)
3. Observe the statusline in any other open session flip to `[VBW] no project | rest-api:<branch>`

## Related

Follows on from #258 / #259. That fix was "walk up to find the config"; this fix is "walk up from the right starting point."

halindrome:
Fixes #266

## Summary

- **`vbw-statusline.sh`**: hoist `_SL_SCRIPT_DIR` before all path resolution; derive `_REPO_ROOT` via `cd "$_SL_SCRIPT_DIR" && git rev-parse --show-toplevel` (script-relative, not CWD-relative); pass `$_SL_SCRIPT_DIR` to `find_vbw_root` as the stable anchor
- **`scripts/lib/vbw-config-root.sh`**: `find_vbw_root` accepts an optional `start_dir` arg — walks up from there when provided, falls back to `pwd` only when called without args (all existing callers unchanged)

## Test plan

- [ ] Open a VBW session in a monorepo root; have an agent `cd` to a sub-package; verify statusline stays anchored to the root project
- [ ] `bash -n scripts/vbw-statusline.sh` — syntax clean
- [ ] `bash -n scripts/lib/vbw-config-root.sh` — syntax clean
- [ ] Existing callers of `find_vbw_root` (no args) — `session-start.sh`, `hook-wrapper.sh`, etc. — unaffected (fallback to `pwd` preserved)


dpearson2699:
@halindrome ci/cd is failing

PR #426, opened 2026-04-15
fix(skills): update CLI syntax to match skills@1.5.0

michaelheichler:
## Summary

`/vbw:skills` and `/vbw:init` embed `npx skills …` commands that match an early/pre-1.0 API of [`vercel-labs/skills`](https://github.com/vercel-labs/skills). Against the current `skills@1.5.0` on npm they fail silently or loudly, so the Skills.sh registry integration is effectively broken end-to-end.

Reproduction with a fresh install:

```
$ npx -y -p skills skill add vercel-labs/skills --skill find-skills -g -y
error: unknown option '--skill'

$ npx -y -p skills skill find python
error: unknown command 'find'
```

`detect-stack.sh` therefore always reports `find_skills_available: false`, the bootstrap step in `/vbw:init` never installs `find-skills`, and every registry search in `/vbw:skills` falls through.

## Fix

- Invoke via `npx -y -p skills skill …` — the package's bin is `skill` (singular); calling it as `skills` collides with npm's own command dispatcher on some setups.
- Drop the removed `--skill <name>` flag; pass the skill as the `<source>` positional. For non-registry skills the full GitHub tree URL works: `https://github.com/vercel-labs/skills/tree/main/skills/<name>`.
- Rename `skills find` → `skill search` (the CLI was renamed).
- Drop `-y` from `skill add` (not a valid flag; kept on `npx` itself).
- Correct the displayed global install path from `~/.agents/skills/` to `~/.claude/skills/`, which is what `skill add -g` actually uses per its own `--help`.

Files touched: `commands/init.md`, `commands/skills.md`.

## Verified

Against `skills@1.5.0`:

```
$ npx -y -p skills skill add \
    https://github.com/vercel-labs/skills/tree/main/skills/find-skills -g
✔ Installed find-skills (global)
  → /home/tux/.claude/skills/find-skills

$ bash scripts/detect-stack.sh . | jq .find_skills_available
true

$ npx -y -p skills skill search python
<registry results>
```

## Test plan

- [ ] /vbw:init on fresh machine: bootstrap installs find-skills without error
- [ ] /vbw:skills: registry search returns results for detected stack
- [ ] /vbw:skills install: both project and global scope succeed
- [ ] detect-stack.sh reports find_skills_available: true after install

### DIFF
--- a/commands/init.md
+++ b/commands/init.md
@@ -243,7 +243,7 @@ If any settings.json changes or plugins installed: display `(restart Claude Code
 **2d. find-skills bootstrap:** Check `find_skills_available` from detect-stack JSON.
 - `true`: display "✓ Skills.sh registry — available"
 - `false`: AskUserQuestion: "○ Skills.sh Registry\n\nVBW can search the Skills.sh registry (~2000 community skills) to find\nskills matching your project. This requires the find-skills meta-skill.\nInstall it now?" Options: "Install (Recommended)" / "Skip"
-  - Approved: `npx skills add vercel-labs/skills --skill find-skills -g -y`
+  - Approved: `npx -y -p skills skill add https://github.com/vercel-labs/skills/tree/main/skills/find-skills -g`
   - Declined: "○ Skipped. Run /vbw:skills later to search the registry."
 
 ### Step 3: Convergence — augment and search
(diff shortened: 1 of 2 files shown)


dpearson2699:
Thanks for digging into the Skills.sh integration — you identified a real issue, but I think the diagnosis landed on the wrong root cause.

**What's correct in the current VBW code:**
- `npx skills add` / `npx skills find` — upstream `skills@1.5.0` `package.json` exposes `bin.skills`, not `bin.skill`
- `--skill` flag — valid per upstream docs
- `-y` / `--yes` — valid per upstream docs

The reproduction commands in the PR used `npx -y -p skills skill ...` (singular `skill`), but the upstream binary is `skills` (plural). So those errors were from invoking a non-existent binary, not from the current VBW syntax being wrong.

**What IS a real bug:**
`commands/skills.md` Step 5b displays `~/.agents/skills/` as the global install path, but `skills@1.5.0` actually installs globals to `~/.claude/skills/` for Claude Code. This is a display text issue — the actual install commands work correctly.

I've filed #433 to track the one real bug and will put up a corrected PR.

Closing this one — the syntax changes would break the CLI integration since `skill` (singular), `search`, and dropping `--skill`/`-y` don't match the upstream API.

PR #639, opened 2026-05-13
fix(statusline): add CLAUDE_CODE_CONTEXT_WINDOW_FLOOR for under-reported context windows

### LINKED ISSUE #638: Status line shows 167K context window for Sonnet 4.6 (should be 1M), causing premature compaction
## Bug

When using `claude-sonnet-4-6` (1M token context window), the VBW status line displays a context window of **~167K tokens** instead of ~1M, and compaction triggers far too early (at ~16% actual usage).

## Root Cause

`vbw-statusline.sh` reads `context_window_size` from the Claude Code hook event payload. Claude Code may report `200000` (or omit the field entirely) for models that actually have 1M context windows. The autocompact normalization then computes the compaction trigger against 200K:

```
effective = 200000 - 20000 = 180000
trigger   = 180000 - 13000 = 167000   ← displayed as the window size
```

`CTX_SIZE` is overwritten with the trigger value (`CTX_SIZE="$_TRIGGER"` at line ~456), so both the displayed window and the `.context-usage` cache used by `suggest-compact.sh` reflect 167K instead of the actual 1M.

## Impact

- Status line shows `167K` context window for Sonnet 4.6, Opus 4.7, Haiku 4.5 (all 1M models)
- Compaction fires at ~167K tokens — only ~16% of the 1M window
- Pre-flight context guard (`suggest-compact.sh`) uses the stale 167K cached value, issuing false warnings

## Fix

Add `config/model-context-windows.json` mapping known model IDs to their correct context windows. After extracting `CTX_SIZE` from the hook event, read `.model.id` from the same payload and apply `max(reported, known)` before the trigger calculation. For `claude-sonnet-4-6` this corrects 200K → 1M, giving a trigger of ~967K.

## Environment

- Model: `claude-sonnet-4-6`
- VBW: v1.37.0
- Symptom: status line shows `XXXk/167k` instead of `XXXk/967k`

halindrome:
## Summary

- Adds `CLAUDE_CODE_CONTEXT_WINDOW_FLOOR` env var (real env or `settings.json .env` block)
- When set, raises `CTX_SIZE` to the floor value before the compaction trigger is calculated
- Reverts the prior incorrect `config/model-context-windows.json` approach
- Updates stale `token-budgets.json` comment about Claude 4.x context windows

Fixes #638

## Why the model-lookup approach was wrong

Context window size for Sonnet 4.6 depends on subscription tier + extra usage:
- All paid plans (chat): 500K
- Claude Code, paid plan, extra usage enabled: 1M
- Claude Code, no extra usage: unknown (likely 200K or 500K)

Hardcoding 1M for all Sonnet 4.6 users would break the status line for users who
genuinely have a smaller window. Claude Code already knows the correct value for
the user's subscription — if it reports 200K while the user has 1M, that is a
Claude Code under-reporting issue.

## The correct fix

A user-configurable floor that the affected user can set explicitly:

```json
// settings.json
{ "env": { "CLAUDE_CODE_CONTEXT_WINDOW_FLOOR": "1000000" } }
```

Same pattern as the existing `CLAUDE_CODE_AUTO_COMPACT_WINDOW` cap, but working
upward instead of downward.

## Test plan

- [ ] Shell syntax: `bash -n scripts/vbw-statusline.sh`
- [ ] ShellCheck: `shellcheck -S warning scripts/vbw-statusline.sh`
- [ ] Contract tests: `bash testing/verify-bash-scripts-contract.sh`
- [ ] Smoke: set `CLAUDE_CODE_CONTEXT_WINDOW_FLOOR=1000000` in settings.json, status line shows ~967K window

### DIFF
--- a/config/model-context-windows.json
+++ b/config/model-context-windows.json
@@ -1,9 +1,0 @@
-{
-  "_comment": "Known context window sizes by model ID. Used to correct under-reported or missing context_window_size values from Claude Code hook events. Update when new large-context models are released.",
-  "windows": {
-    "claude-opus-4-7": 1000000,
-    "claude-sonnet-4-6": 1000000,
-    "claude-haiku-4-5": 1000000,
-    "claude-haiku-4-5-20251001": 1000000
-  }
-}
--- a/config/model-context-windows.json
+++ b/config/model-context-windows.json
@@ -1,0 +1,9 @@
+{
+  "_comment": "Known context window sizes by model ID. Used to correct under-reported or missing context_window_size values from Claude Code hook events. Update when new large-context models are released.",
+  "windows": {
+    "claude-opus-4-7": 1000000,
+    "claude-sonnet-4-6": 1000000,
+    "claude-haiku-4-5": 1000000,
+    "claude-haiku-4-5-20251001": 1000000
+  }
+}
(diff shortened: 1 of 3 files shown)


halindrome:
This is NOT a good idea.

PR #641, opened 2026-05-20
fix(planning-root): submodule-aware planning discovery

### LINKED ISSUE #640: VBW commands fail when cwd is inside a git submodule of the monorepo
**Command**
/vbw:status (and any VCS/planning operation that consults `.vbw-planning/`)

**What happened**
In a git-submodule-based monorepo, an earlier operation can leave the working directory inside a submodule. Subsequent VBW commands (e.g. `/vbw:status`) then look for `.vbw-planning/ROADMAP.md` relative to the current submodule directory, fail the "No roadmap found" guard, and stop — even though the parent workspace at the monorepo root does have a valid `.vbw-planning/` directory.

Concrete example from a Corvex monorepo at `/Users/ahby/Sources/liverpool` with a submodule at `mcp-server/`:

```
~/Sources/liverpool/mcp-server  ❯ /vbw:status
⏺ The status reflects the submodule's state (no .vbw-planning/ROADMAP.md in mcp-server/).
  Per the guard: No roadmap found.
```

Running the same command from `~/Sources/liverpool/` (the monorepo root) finds the planning directory and proceeds normally.

**What you expected**
VBW commands should walk up from the current directory to find the nearest ancestor that contains `.vbw-planning/` — the same way `git` discovers the repo root via the `.git` directory. Submodules of a VBW-managed monorepo should inherit the parent's planning context. At minimum, the guard message should detect the situation and emit a clearer hint such as "VBW planning found at `<parent>` — re-run from there" rather than the generic "No roadmap found".

**Steps to reproduce**
1. Have a monorepo with `.vbw-planning/` at the root and one or more git submodules.
2. From the monorepo root, `cd` into a submodule (or have a tool/agent do so during execution).
3. Run `/vbw:status` (or any VBW command that consults the planning dir).
4. Observe the "No roadmap found" guard despite valid planning at the parent.

**Environment**
- Claude Code version: 2.1.143 (Claude Code)
- OS: Darwin 25.2.0 arm64
- Plugin install method: marketplace
- VBW version: 1.37.1
- Model: Not specified

**Additional context**

halindrome:
## Summary

Fixes the bug where VBW commands trip `Run /vbw:init first` when CWD is a submodule subdirectory of a workspace that has `.vbw-planning/` at the parent.

Phase 11 of the local roadmap; ships as a single PR per maintainer preference (originally scoped as three sequential PRs, combined for one review cycle).

## Why

Reported in #640. Before this PR, VBW commands resolved `.vbw-planning/` only from the literal CWD. Anyone running VBW from a git submodule subdirectory or a sibling package directory got a false "not initialized" message even though the workspace planning state lived one or two directories up.

## How

Three logical layers (P1/P2/P3):

**P1 — Resolver bridge + library cascade + bare-reference migration**
- New `scripts/resolve-planning-root.sh` stdout wrapper for command precompute blocks.
- `scripts/lib/vbw-config-root.sh` implements a 4-step cascade in `find_vbw_root`:
  1. `VBW_PLANNING_ROOT` env var override
  2. `$(git rev-parse --git-common-dir)/info/vbw-planning-root.txt` pointer file
  3. Filesystem walk-up from CWD until ancestor `.vbw-planning/config.json` found
  4. CWD fallback (now emits a deprecation banner)
- One-line stderr auto-resolve banner emitted once per shell process when CWD maps to an ancestor planning root.
- 22 bare-reference scripts source the lib and call `find_vbw_root` before reading `.vbw-planning/`.
- 19 command precompute blocks route through `resolve-planning-root.sh`.
- `AGENTS.md` documents the pointer-file convention (mirrors the existing `vbw-debug-target.txt` precedent).

**P2 — Command runtime-shell migration**
- ~80+ runtime shell references in 12 command files (`vibe.md`, `config.md`, `debug.md`, `qa.md`, `verify.md`, `resume.md`, `fix.md`, `status.md`, `map.md`, `pause.md`, `discuss.md`, `todo.md`) rewritten from bare `.vbw-planning/` to `$PLANNING_ROOT`.
- 16 guard messages converted to a three-branch pattern (truly-uninitialized stop / ancestor-found banner-and-continue / root-at-cwd continue).
- `scripts/delegated-workflow.sh` migrated to source the lib and use the three-branch guard.

**P3 — Deprecation banner + bats coverage**
- `find_vbw_root` emits a single stderr deprecation banner (sentinel-guarded, once per shell) when the cascade falls through to the CWD-only fallback.
- `tests/submodule-planning-discovery.bats` adds 5 new scenarios: CWD at planning root, CWD in non-submodule subdir, CWD inside a real git submodule, CWD in a sibling git worktree using the pointer file, CWD with no ancestor planning root (banner emitted).
- One additional library-layer pointer-file unit test.

## Test plan

- [x] `bash testing/run-all.sh` — lint clean, contract clean, BATS 3596 passed, 1 failed
- [x] Sole failing test is `tests/role-isolation-runtime.bats` "scout fails open" — reproduces on `main` pre-this-PR (pre-existing, out of scope)
- [x] Hands-on submodule repro: from a subdir of a workspace with `.vbw-planning/`, VBW commands now resolve correctly; "Run /vbw:init first" no longer false-trips
- [x] Local UAT walkthrough complete (13 checkpoints, 12 pass / 1 skip / 0 issues)

## Files

- 42 files changed, +662 / −186
- New: `scripts/resolve-planning-root.sh`, `scripts/lib/vbw-config-root.sh`, `tests/submodule-planning-discovery.bats`, `tests/planning-git-callsites.bats`, `testing/verify-plugin-root-resolution.sh`
- Modified: 12 commands, 22 bare-reference scripts (sourcing the lib), `AGENTS.md`, `scripts/delegated-workflow.sh`, `scripts/phase-detect.sh`, two `verify-*.sh` assertion files

## Notes for reviewers

- `scripts/phase-detect.sh` uses a `VBW_PLANNING_DIR` snapshot/restore wrapper around `find_vbw_root "$SCRIPT_DIR"` to preserve CWD-as-workspace output semantics that 8 `verify-plan-filename-convention.sh` fixtures depend on. Otherwise the resolver fallback would pull in the host repo's planning root when CWD is a test fixture without a `config.json`.
- The auto-resolve banner and deprecation banner use separate sentinel guards so a process that legitimately encounters both does not silently suppress one.
- Pointer file lives under `$(git rev-parse --git-common-dir)/info/` so it stays private (not tracked) and is shared across all worktrees of the same clone.

Fixes #640


halindrome:
## QA Round 1

```markdown
# PR #641 QA Review — fix/640-p3-fallback-bats

- **Model used:** Claude Opus 4.7
- **Scope:** Resolver core, sentinel guards, phase-detect snapshot/restore, command precompute call sites, three-branch guard pattern, new bats tests.

## Findings

### CONFIRMED — Major

1. **VBW_PLANNING_ROOT env var is not validated and bypasses pointer file + walk-up**
   - scripts/lib/vbw-config-root.sh:114-119 and scripts/resolve-planning-root.sh:22-25 accept any non-empty value and export it verbatim. No `[ -d "$VBW_PLANNING_ROOT" ]` test (compare to pointer-file step at line 129). Stale/typo env var silently produces a non-existent path; user gets misleading "Run /vbw:init first."
   - **Severity: Major.** Confirmed by reading code.

2. **resolve-planning-root.sh early-exit on VBW_PLANNING_ROOT prints the value raw — but find_vbw_root exports VBW_PLANNING_DIR="$VBW_PLANNING_ROOT/.vbw-planning"**
   - resolve-planning-root.sh:23 prints `$VBW_PLANNING_ROOT` (the workspace root). Downstream consumers expect the `.vbw-planning` directory. Contract mismatch between early-exit and post-find_vbw_root branches (line 119 prints `$VBW_PLANNING_DIR` with the suffix).
   - When VBW_PLANNING_ROOT=/abs/path is set, the bridge prints /abs/path; consumer does "$PLANNING_ROOT/config.json" → /abs/path/config.json (wrong).
   - **Severity: Critical** if anyone uses VBW_PLANNING_ROOT as documented. Confirmed.

3. **list-todos.md was not migrated**
   - commands/list-todos.md:20 still says STOP "Run /vbw:init first." Submodule subdirs running /vbw:list-todos will still hit the regression #640 set out to fix.
   - **Severity: Major.** Confirmed.

4. **teach.md and profile.md precompute uses bridge but guard remains literal**
   - teach.md:30 and profile.md:21 still reference bare `.vbw-planning/` in the guard prose even though precompute resolves the right path. Instruction text is self-contradictory in submodule cwd.
   - **Severity: Major.**

### CONFIRMED — Minor

5. Three-branch guard middle branch uses "is an ancestor of cwd" (English) but truly-uninit branch uses string equality on dirname — fuzzy contract.
6. Pointer file: no rejection of relative path (CLAUDE.md says relative paths are rejected).
7. Sentinel collisions across nested shells (`_VBW_BANNER_EMITTED` / `_VBW_FALLBACK_BANNER_EMITTED` exported; persists across nested bash).
8. uninstall.md:45 still uses `rm -rf .vbw-planning/` (bare literal).

### HYPOTHETICAL — Needs runtime verification

9. debug.md has 24 bare `.vbw-planning` refs in instruction-string literals for subagent prompts. Migration appears incomplete. Severity: Major-hypothetical.
10. vibe.md has 48 bare `.vbw-planning` refs in instruction strings. Spot-check shows runtime shell calls in spawned-agent prompt text. Severity: Major-hypothetical.
11. **Phase-detect snapshot/restore unconditionally restores VBW_PLANNING_DIR** — when phase-detect is invoked from a submodule subdir with no env pre-set, VBW_PLANNING_DIR is restored to empty and `PLANNING_DIR="${VBW_PLANNING_DIR:-.vbw-planning}"` falls to literal `.vbw-planning` — breaking the #640 fix for phase-detect callers. Needs bats coverage of phase-detect-from-submodule. Severity: Critical-hypothetical.
12. role-isolation-runtime.bats "scout fails open" punted as pre-existing — violates CLAUDE.md zero-tolerance test policy. Severity: Major (process violation).

### Test coverage gaps

13. New bats covers T1-T5. Not covered: VBW_PLANNING_ROOT env var path, malformed pointer file (relative path, multiple lines, leading whitespace, non-existent target, file-not-dir), sentinel idempotency, phase-detect.sh from submodule subdir.

## Summary

- **Critical**: #2 (env-var bridge contract mismatch), #11 (phase-detect snapshot may break fix in submodule mode — needs runtime check)
- **Major**: #1, #3, #4, #9, #10, #12
- **Minor**: #5, #6, #7, #8
```

halindrome:
## QA Round 2

```markdown
# QA Round 2 Report — PR #641 (fix/640-p3-fallback-bats)

Model used: Claude Opus 4.7
Scope: Verify Round 1 fixes (commit 36bf0efe) and hunt for newly-introduced or overlooked issues.

## Round 1 fixes — verified

All 5 Round 1 fixes confirmed by reading code: env-var early-exit, validation, pointer-file relative rejection, phase-detect conditional restore, four command guards migrated.

## New / overlooked findings

### R2-1 — Major (confirmed): phase-detect "real ancestor" check too strict

`scripts/phase-detect.sh:60-68` gates the resolver-result-keep branch on `[ -f "${VBW_PLANNING_DIR}/config.json" ]`. In the bootstrap window — workspace where `/vbw:init` has created `.vbw-planning/` but not yet written `config.json` (or a partially-corrupted project) — `find_vbw_root` correctly exports ancestor `VBW_PLANNING_DIR`, but phase-detect *restores* the snapshot, clobbering the resolver result. Recommend `[ -d "${VBW_PLANNING_DIR}" ]` instead — the cascade already guards bogus values.

### R2-2 — Major (confirmed): VBW_PLANNING_ROOT validation only checks workspace, not `.vbw-planning/` inside

`scripts/lib/vbw-config-root.sh:118-122` validates `$VBW_PLANNING_ROOT` is a dir but not that `$VBW_PLANNING_ROOT/.vbw-planning/` exists. User pointing env var at a non-VBW workspace gets the misleading "Run /vbw:init first" UX rather than a clear "env var misconfigured" signal. Tighten to `[ -d "$VBW_PLANNING_ROOT/.vbw-planning" ]` or emit a stderr diagnostic.

### R2-3 — Minor (confirmed): pointer-file relative-path rejection is silent

`scripts/lib/vbw-config-root.sh:135-138` drops relative values with no stderr emission. Suggest one-line warning.

### R2-4 — Major (confirmed): 11 commands still use bare-text init guard

Round 1 migrated only `list-todos`, `teach`, `profile`, `uninstall`. The other 11 commands matched by `grep "Run /vbw:init first"` still use bare `.vbw-planning/` guard prose: `debug.md`, `pause.md`, `fix.md`, `map.md`, `qa.md`, `status.md`, `verify.md`, `todo.md`, `discuss.md`, `resume.md`, `config.md`. Submodule callers will still hit the original #640 bug from those commands. (`vibe.md` is already migrated via the precompute key pattern.)

This is the very bug class #640 set out to fix.

### R2-5 — Minor (confirmed): no new bats coverage for Round 1 contracts

`tests/submodule-planning-discovery.bats` covers the original 5 scenarios but adds no tests for the 4 new Round 1 contracts: env-var honored when valid, env-var falls through when stale, pointer file rejected when relative, phase-detect ancestor preservation. CI green is no-regression, not coverage.

### R2-6 — Hypothetical: scripts/state-updater.sh:366 hardcodes `echo ".vbw-planning"`

Worth a 30-second look before merge — if this is returning a relative path callers concatenate, it bypasses the cascade.

## Cumulative interaction

R2-1, R2-2, and R2-4 all funnel users into the same dead-end UX ("Run /vbw:init first") even when the workspace is initialized. Round 1 fixed 4 commands; 11 remain.

## Recommendation

- **Block merge on R2-4** (eleven commands not migrated).
- Tighten R2-1 (`-d "$VBW_PLANNING_DIR"`).
- Add at least one bats covering env-var honored + falls-through, and one covering phase-detect ancestor preservation (R2-5).
- R2-2, R2-3, R2-6 non-blocking but worth tracking.
```

halindrome:
## QA Round 3

```markdown
# QA Round 3 Report — PR #641 (fix/640-p3-fallback-bats)

Model used: Claude Opus 4.7
Scope: cumulative R1 + R2 fix surface (commits 36bf0efe + 6a5d6709). Hunt for new regressions and edge cases introduced by the fixes.

## Findings

### R3-1 — Major (confirmed): uninstall.md references $PLANNING_ROOT but never Binds it

`commands/uninstall.md:45` uses `"$PLANNING_ROOT"` and `rm -rf "$PLANNING_ROOT"` but no Bind directive (`grep -n PLANNING_ROOT commands/uninstall.md` returns exactly one line — the consumer). The other three migrated commands (`list-todos`, `teach`, `profile`) all have explicit Bind lines. The precompute block at uninstall.md:20 uses a LOCAL `P=$(bash ...)` inside `!` execution scope which does not survive.

Impact: when the LLM runs `rm -rf "$PLANNING_ROOT"`, `$PLANNING_ROOT` is empty — destructive command with missing operand. Worst case the LLM helpfully substitutes the bare `.vbw-planning/` literal it sees in the diagnostic, which contradicts the new prose explicitly forbidding it.

**Severity: Major. Block on this.**

### R3-2 — Minor (confirmed): symlink realpath mismatch breaks ancestor case-match

`scripts/lib/vbw-config-root.sh:128` (env-var branch) and `:157` (pointer-file branch) set `VBW_CONFIG_ROOT` verbatim. Walk-up branches use `pwd -P`-derived values. `scripts/phase-detect.sh:59` computes `_pd_cwd_now=$(pwd -P)` then matches `case "$_pd_cwd_now/" in "$VBW_CONFIG_ROOT"/*)`. If user sets `VBW_PLANNING_ROOT=/symlinked/path` and runs phase-detect from `/real/path/sub`, case fails — silent drop. Fix: realpath-resolve `VBW_CONFIG_ROOT` at env-var + pointer-file branches.

### R3-3 — Minor (hypothetical): resolve-planning-root.sh env-var early-exit doesn't export VBW_CONFIG_ROOT

Bridge env-var branch exits before sourcing the lib. Subprocess callers don't care (exports wouldn't propagate anyway), but any helper that sources the bridge would skip the lib entirely. No such caller found today.

### Clean areas

- Sentinel names don't collide (4 distinct `_VBW_*` names).
- Diagnostic stderr does not leak into /vbw:status format-sensitive output (all resolver invocations in status.md use `2>/dev/null`).
- Bash 3.2 / macOS portability: clean.
- Three-branch guard prose vs lib `case` pattern: informal but accurate (deferred from Round 1, no functional defect).

## Recommendation

- Block on R3-1 (one-line Bind directive in uninstall.md).
- R3-2 worth fixing — realpath-normalize at env-var and pointer-file branches.
- R3-3 deferred.
```

halindrome:
## QA Round 4

```markdown
# QA Round 4 Report — PR #641 (fix/640-p3-fallback-bats)

Model used: Claude Opus 4.7
Scope: verify R3 fixes (commit 134e2e82) and hunt for remaining issues.

## Findings

### R3-1 (uninstall.md Bind) — CONFIRMED CLEAN

commands/uninstall.md:31 adds an explicit Bind directive before Step 1 with NON-NEGOTIABLE language and `|| echo .vbw-planning` fallback. Placement and wording match list-todos/teach/profile convention. The LLM encounters it before Step 4's destructive `rm -rf "$PLANNING_ROOT"`. All 16 commands referencing `$PLANNING_ROOT` now have a corresponding Bind/initialization directive (audited by grep across all commands/*.md).

### R3-2 (realpath normalization) — CONFIRMED CLEAN

scripts/lib/vbw-config-root.sh:128-131 (env-var branch) and :161-164 (pointer-file branch) use `_vbw_X_canon=$(cd "$X" 2>/dev/null && pwd -P 2>/dev/null) || _vbw_X_canon="$X"`. Set-u safe (unconditional assignment via `|| fallback`); 14 callers with `set -u` cannot trip nounset. No shell-quoting hazards (both inputs double-quoted).

### R4-1 (Minor, confirmed): no bats coverage for symlinked env-var/pointer-file paths

`git diff origin/main..HEAD -- testing/**/*.bats` returns zero files matching `symlink|ln -s|readlink|realpath|pwd -P`. R3-2's realpath fix is correct by inspection but lacks a regression guard. Suggest a tmpdir-based bats that sets `VBW_PLANNING_ROOT=/tmp/symlink-to-workspace` and asserts `VBW_CONFIG_ROOT` equals the realpath target. Non-blocking.

## Verdict

**CLEAN for behavior.** One Minor test-coverage observation (R4-1), which is non-blocking under CONTRIBUTING.md's "clean or only hypothetical/minor findings" gate.

R3-1 + R3-2 correctly implemented; no remaining `$PLANNING_ROOT` references lack binding; no shell-quoting/set-u issues introduced; commit body claims accurate.
```

halindrome:
## QA Round 5

```markdown
# QA Round 5 Report — PR #641 (fix/640-p3-fallback-bats)

Model used: Claude Opus 4.7
Scope: investigate and fix two Linux-only CI failures present since the original PR-open commit (cb6f7687). origin/main was green; these failures are introduced by Phase 11's P1-P3 lib changes.

## Failures investigated

1. `tests/statusline-cache-isolation.bats:206-209` — "repo with remote shows GitHub link, not bare directory name". Failure: `head -1` of `bash $STATUSLINE 2>&1` did not contain `]8;;https://`.
2. `tests/statusline-hide-limits.bats:189-219` — "statusline_collapse_agent_in_tmux: collapses to single line in git worktree". Failure: first line of statusline output did not contain "Model:".

## Root cause

Phase 11's P3 work in `scripts/lib/vbw-config-root.sh` added user-facing stderr banners (`_emit_vbw_fallback_banner`, `_emit_vbw_auto_resolve_banner`) and QA Round 2 added two more diagnostic warnings (env-var no-planning, pointer-file relative path).

The failing tests use `bash $STATUSLINE 2>&1 | head -1`. When statusline runs in any environment where `find_vbw_root "$_SL_SCRIPT_DIR"` cannot resolve a planning root (e.g., a Linux CI runner where `.vbw-planning/` is gitignored and absent in the checkout), the fallback banner fires on stderr. Merged via `2>&1`, the banner becomes line 1 and the OSC 8 / "Model:" assertions fail. On macOS this branch had a stale local `.vbw-planning/config.json` so walk-up succeeded and no banner fired.

This is a real production bug, not just a test artifact: every statusline render on a clean clone would dump a warning into the terminal pane.

## Fix (root-cause, two minimal edits)

1. **`scripts/lib/vbw-config-root.sh`**: gate all four user-facing diagnostics behind `VBW_SUPPRESS_BANNERS=1`. The two banner emitters and two warning paths each early-return when the var is set. Bats tests that ASSERT banner emission (T5/T8/T9 in submodule-planning-discovery.bats) continue to pass because they don't set the var.

2. **`scripts/vbw-statusline.sh`**: `export VBW_SUPPRESS_BANNERS=1` before calling `find_vbw_root`. The statusline is a continuously-rendered terminal artifact; stderr noise there is always undesired (both in tests and in production).

## Verification

- 220 bats PASS across submodule-planning-discovery, statusline-cache-isolation, statusline-hide-limits, phase-detect, planning-git-callsites, monorepo-hook-cwd
- Contract scripts clean (init-todo, commands-contract, bash-scripts-contract, plugin-root-resolution)
- Lint clean
- Both previously-failing tests now pass locally with the fix
```

scripts/planning-git.sh
```diff
@@ -168,8 +169,9 @@ case "$COMMAND" in
 
     ensure_transient_ignore
 
-    if [ -d ".vbw-planning" ]; then
-      git add .vbw-planning
+    _pg_planning_dir="${VBW_PLANNING_DIR:-.vbw-planning}"
```
dpearson2699, line 201: `VBW-PR-001`: This still runs Git from the caller's CWD. From inside a real submodule, the parent planning dir is outside the submodule repo, so `git add "$_pg_planning_dir"` fails; without `VBW_PLANNING_DIR` it silently targets `.vbw-planning` in the submodule. Passing `"$PLANNING_ROOT/config.json"` from commands is not enough while `planning-git.sh` remains rooted at the submodule.

commands/list-todos.md
```diff
@@ -17,7 +17,12 @@ allowed-tools: Read, Edit, Bash
 
 ## Guard
 
-1. **Not initialized** (no .vbw-planning/ dir): STOP "Run /vbw:init first."
+Bind `PLANNING_ROOT=$(bash "/tmp/.vbw-plugin-root-link-${CLAUDE_SESSION_ID:-default}/scripts/resolve-planning-root.sh" 2>/dev/null || echo .vbw-planning)` and use `"$PLANNING_ROOT/..."` for every subsequent path reference in this command.
```
dpearson2699: `VBW-PR-002`: This captures `PLANNING_ROOT`, but the helper subprocess below does not receive it and `todo-lifecycle.sh`/`list-todos.sh` default to `VBW_PLANNING_DIR:-.vbw-planning`. From a submodule CWD this still reads under the submodule and returns `STATE.md not found` even though the resolver found the parent planning dir. Pass or export the resolved planning dir before invoking helpers that still own `.vbw-planning` lookup.

commands/status.md
```diff
@@ -101,17 +101,21 @@ fi`
...
 Shipped milestones:
 ```
-!`ls -d .vbw-planning/milestones/*/SHIPPED.md 2>/dev/null || echo "No shipped milestones"`
+!`P=$(bash "{plugin-root}/scripts/resolve-planning-root.sh" 2>/dev/null || echo ".vbw-planning"); ls -d "$P"/milestones/*/SHIPPED.md 2>/dev/null || echo "No shipped milestones"`
 ```
 
 ## Guard
 
-- Not initialized (no .vbw-planning/ dir): STOP "Run /vbw:init first."
+Bind `PLANNING_ROOT=$(bash "/tmp/.vbw-plugin-root-link-${CLAUDE_SESSION_ID:-default}/scripts/resolve-planning-root.sh" 2>/dev/null || echo .vbw-planning)` and use `"$PLANNING_ROOT/..."` for every subsequent path reference in this command.
```
dpearson2699, line 109: `VBW-PR-003`: This new guard is contradicted by later status steps that still instruct bare `.vbw-planning/phases/` and `.vbw-planning/milestones/` reads. From a submodule CWD those steps revert to submodule-relative paths even though the resolver found the parent workspace. The remaining bare runtime path instructions need to be migrated to the resolved planning root.

dpearson2699 (changes requested):
Requesting changes. The resolver/library layer is directionally right, but the PR does not yet satisfy issue #640 for commands executed from submodule CWDs.

Blocking findings are inline:

- `VBW-PR-001`: `planning-git.sh` still runs Git from the caller's CWD/repo, so parent planning state is not handled correctly from a real submodule.
- `VBW-PR-002`: command helpers are invoked without passing/exporting the resolved planning dir, so they still default to cwd-relative `.vbw-planning`.
- `VBW-PR-003`: command templates still instruct later runtime reads/writes through bare `.vbw-planning` paths after the new `PLANNING_ROOT` guard.

Unanchored blockers: none.

Verification:
- Compared against linked issue #640 and the blind baseline expectation that planning and VCS consumers use the resolved parent planning root.
- QA investigator reproduced the helper/planning-git failures with targeted probes.
- Targeted PR tests passed: `bats tests/submodule-planning-discovery.bats tests/monorepo-hook-cwd.bats tests/planning-git-callsites.bats` (46/46).
- Local full `bash testing/run-all.sh` did not complete: it hung in `tests/agent-health.bats` on `agent-health: orphan recovery preserves role-owned task when another same-role teammate is still alive` and was interrupted after more than 5 minutes (exit 130). The single-case replay passed on both `main` and this PR branch; GitHub CI/QA/Linked Issue checks are green for `c588b6e`.

halindrome:
Addressed the review feedback in `7f1218dc`. The resolver layer was correct; the three flagged consumers still defaulted to the caller's cwd-relative `.vbw-planning`, so from a submodule CWD they operated on the submodule instead of the parent workspace.

**VBW-PR-001 — `scripts/planning-git.sh` rooting.** It ran git from the caller's CWD, so `git add .vbw-planning` from a submodule targeted the wrong repo / a path outside the submodule work tree. Added submodule-safe rooting before the command dispatch: resolve the planning dir (from `VBW_PLANNING_DIR` or `resolve-planning-root.sh`), `cd` to the workspace root that owns it, and pin `VBW_PLANNING_DIR` to its absolute path — so `sync-ignore`, `commit-boundary`, and `push-after-phase` all act on the workspace repo. (Passing `$PLANNING_ROOT` from commands alone could not fix this while the script stayed rooted at the caller's CWD.)

**VBW-PR-002 — `commands/list-todos.md` helper threading.** Each Bash block is its own subprocess, so capturing `PLANNING_ROOT` wasn't enough — `todo-lifecycle.sh`/`suggest-next.sh` fell back to `VBW_PLANNING_DIR:-.vbw-planning` and reported `STATE.md not found`. Every helper invocation now carries `VBW_PLANNING_DIR="$PLANNING_ROOT"`, and the guard documents the requirement.

**VBW-PR-003 — `commands/status.md` bare paths.** Migrated the later bare `.vbw-planning/phases`, `.vbw-planning/milestones`, `config.json`, and `.cost-ledger.json` instructions to `"$PLANNING_ROOT/..."`.

Added a behavioral regression test in `tests/submodule-planning-discovery.bats` that builds a real submodule and proves `planning-git.sh commit-boundary` commits to the **outer** workspace repo from a submodule CWD.

`bash testing/run-all.sh` is green (lint 1/1, contracts 51/51, BATS 3606/0; the targeted set `submodule-planning-discovery` / `monorepo-hook-cwd` / `planning-git-callsites` is 46/46 plus the new case). Rebased onto the latest branch tip before pushing.

